PR #36, opened 2018-04-04
Configured XML documentation output

### LINKED ISSUE #35: Include XML Documentation in NuGet Package
Visual Studio allows to make Documentation in Code:
```cs
/// <summary>
/// This class provides an interface to the Raspberry Pi camera module. 
/// </summary>
public sealed class MMALCamera 
{
}
```
But this documentation is not shown as long as you don't generate an XML file at build and include this file in your NuGet package. In Visual Studio you can go to `Project Settings->Build` and enable `XML documentation file`. This will generate a `MMALSharp.xml` in your output directory which you can include to your NuGet package. After this all developers that are using your library will see your documentation in Visual Studio.

daniel-lerch:
I configured the output of XML documentation files. The compiler will now create an XML file in the output directory which resolves #35 .

The `<Service Include="" />` tag in ` tests/MMALSharp.Tests/MMALSharp.Tests.csproj` was automatically made by Visual Studio in order to optimize startup performance with XUnit ([see stackoverflow](https://stackoverflow.com/questions/18614342/what-is-service-include-in-a-csproj-file-for)).

### DIFF
--- a/src/MMALSharp.Common/MMALSharp.Common.csproj
+++ b/src/MMALSharp.Common/MMALSharp.Common.csproj
@@ -1,25 +1,31 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">  
+﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\build\MMALSharp.Common.props" />
   <PropertyGroup>
     <AssemblyTitle>MMALSharp.Common</AssemblyTitle>
-    <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>    
+    <TargetFrameworks>net452;netstandard2.0</TargetFrameworks>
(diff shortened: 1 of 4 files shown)


daniel-lerch:
Maybe we should suppress warning `CS1591 "Missing XML comment for publicy visible type or member"` because we have 1830 warnings and the AppVeyor output is completely boated by warning messages.

techyian:
Thanks for this Daniel. I'm in two minds about suppressing the warnings - it would be a good exercise to try and work through the remaining Managed methods that are missing XML documentation. I can see the majority however are coming from the native areas so I'm happy to suppress in these files.

techyian:
Added #37 for the remaining areas and suppress in native areas.

Issue #51, opened 2018-05-13
ResizerComponent.ConfigureOutputPort doesn't work


daniel-lerch:
Hi Ian,

I am still trying to get an uncompressed image from MMALSharp.
At first I tried your approach from `TakeRawPicture` but got an EINVAL when I tried to commit the image format ARGB. Then I searched with the new utility method from #17 for available encodings.
Finally I tried many versions with ResizerComponent all resulting in an EINVAL at ConfigureOutputPort. No matter whether I use `I420`, `BGRA` or `YUYV`, it always crashes.

Here is my code:
```csharp
public int Width => Resolution.As8MPixel.Width;
public int Height => Resolution.As8MPixel.Height;

public void Initialize()
{
    MemoryCaptureHandler stillHandler = new MemoryCaptureHandler(null);
    MMALResizerComponent stillResizer = new MMALResizerComponent(Width, Height, stillHandler);
    MMALNullSinkComponent previewSink = new MMALNullSinkComponent();

    MMALCameraConfig.StillResolution = Resolution.As8MPixel;
    cam.ConfigureCameraSettings();

    PrintEncoding(cam.Camera.StillPort.GetSupportedEncodings(), "Still Port");
    PrintEncoding(stillResizer.Outputs[0].GetSupportedEncodings(), "Resizer");

    stillResizer.ConfigureInputPort(MMALEncoding.OPAQUE, MMALEncoding.I420, cam.Camera.StillPort);
    stillResizer.ConfigureOutputPort(MMALEncoding.I420, MMALEncoding.I420, 0);
    cam.Camera.PreviewPort.ConnectTo(previewSink);
    cam.Camera.StillPort.ConnectTo(stillResizer);
}
```
The `MemoryCaptureHandler` simply writes all bytes on a MemoryStream.

Daniel

techyian:
Strange. Seems to be working fine for me under Mono 4.6.2. My code is fairly similar to yours except is wrapped in `AsyncContext.Run()`, although that wouldn't cause the EINVAL. Which runtime are you using with your code? Also, which Pi and Camera Module are you using?

daniel-lerch:
I am using a Raspberry Pi 3B with the Sony IMX219 camera module.
On Raspbian I am running an [Avalonia](https://github.com/AvaloniaUI/Avalonia) Application which should provide its own async context. But the application crashes before calling any async method.
The application is built on .NET Core 2.0 and published in Debug configuration as a self-contained app for _linux-arm_.

[...]

Issue #37, opened 2018-04-04
Update XML documentation


techyian:
Update XML documentation for currently implemented Managed areas of MMALSharp.

Suppress warnings in native areas.

daniel-lerch:
Hi Ian,

I have started to add XML documentation to the code. Because I first came in touch with MMAL five days ago I don't understand all your code yet but I'll give my best to contribute to the documentation.

How requently do you want to get pull requests?

Daniel

techyian:
Hi Daniel,

Thank you for getting involved in the project. I would encourage you to look into the Native MMAL API that the Raspberry Pi team maintain first if you're new to it all. 

The native API docs can be found [here](http://www.jvcref.com/files/PI/documentation/html/) and help with understanding what each of the native calls do.

For more theory into the actual workings of MMAL and the Raspberry Pi camera itself, the PiCamera python library has extensive documentation [here](http://picamera.readthedocs.io/en/latest/). I'd certainly like this project to have that kind of documentation in the future, but as I'm sure you can appreciate, it all takes up a lot of time.

There's quite a lot to the API itself, but it will give you a good footing to be able to contribute to MMALSharp.

Ian.

PR #53, opened 2018-05-20
Work on many issues

daniel-lerch:
Hi Ian,

like with all pull requests I added more documentation ( #37 ).
In addition to that I set more restrictive access modifiers in some cases.
Furthermore I added a parameter checking for `MMALResizerComponent` ( #51 ).

In `GenericExtensions.cs` I changed the writing of `float` and `integer` values to make it more obvious which multiplication operator we are using. And I changed the `float.ToByte()` method by multiplying with 255 instead of 256. This is the mathematically right solution if you have a proportional scale from 0 to 255.

Daniel

### DIFF
--- a/src/MMALSharp.Common/GenericExtensions.cs
+++ b/src/MMALSharp.Common/GenericExtensions.cs
@@ -7,8 +7,19 @@ using System;
 
 namespace MMALSharp
 {
+    /// <summary>
+    /// Provides extension methods for mathematical operations.
+    /// </summary>
     public static class GenericExtensions
     {
+        /// <summary>
+        /// Returns a representation of this object that is in the specified range. Too large values will be dreceased to max; too small values will be increased to min.
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="val"></param>
+        /// <param name="min">The mininum inclusive value.</param>
+        /// <param name="max">The maximum inclusive value.</param>
+        /// <returns></returns>
         public static T Clamp<T>(this T val, T min, T max) where T : IComparable<T>
         {
             if (val.CompareTo(min) < 0)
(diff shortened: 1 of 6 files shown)


techyian:
Hi Daniel,

Just want to discuss a few things before accepting the PR. 

1) Multiplying by 255 instead of 256 - my initial thought was to multiply by 255. However, this [answer](https://stackoverflow.com/questions/1914115/converting-color-value-from-float-0-1-to-byte-0-255) on SO suggested 256 is a safe value to multiply by so long as the input param isn't 1.0f, and if so, to handle that separately. Happy to discuss this further.

2) The ranges you've specified in the Resizer Component - are these values referenced anywhere officially, or are they based on your own findings?

Thanks,
Ian

daniel-lerch:
Hi Ian,

> 1. Multiplying by 255 instead of 256...

Sorry about that. I was not aware that the conversion from a float to byte is more like `Math.Floor` than `Math.Round`.
In my opinion using `(byte)Math.Round(val * 255.0f)` would have the least precision loss. However using `(byte)(val * 255.999)` will run slightly faster because you have less instructions and no branch like in current method. Overflow checks also cost performance and are unnecessary because you can't damage more than your RGB values.

> 2. The ranges you've specified in the...

I did not search for official documentation for this issue. I just tried out resolutions. My boundary checks only prevent an EINVAL. Using 16384×16384 crashed my application with an ENOMEM but may work with more RAM.
However there a few other strange results. Using 1×1 as resolution kind of crashed the GPU. It showed coloured stripes all over the display. I could only remove the power supply to reboot.
The next interesting point is that you always receive at least 2048 bytes with `BGRA` encoding. This means resulting images have a resolution of at least 32×16 pixel even if the resolution is configured to 2×2.
The last unexpected behavior was achieved using 32767×32767 as desired resolution. In this case the application did not crash with an ENOMEM but returned no data.

Maybe you can test these settings with your Pi, too. After this I would rather add tips to the wiki than forcing lower/higher values by Exceptions. 
And it would be great if you could add to the wiki that the component is initialized when calling `ConfigureOutputPort`. This would make it much easier to find errors coming from wrong constructor values, etc.

Daniel

daniel-lerch:
Hi Ian,

which approach do you prefer for the colour conversions?
Do you want me to search for an official documentation for supported resolutions with the _ResizerComponent_?
Are there any other points that need to be discussed/changed before you can merge this pull request?

Daniel

PR #156, opened 2020-07-12
migrate FFmpegCaptureHandler to ExternalProcessCaptureHandler, add VL…

MV10:
Project `MMALSharp.Processing`:
- adds `ExternalProcessCaptureHandler`
- async `Channel<T>`-based stdout/stderr buffering and output
- Unix signal-based graceful process termination capabilities
- requires awaiting both `ProcessAsync` and `ManageProcessLifecycleAsync`

As mentioned in the original issue thread, buffering avoids frame drops and other problems due to `Console.WriteLine` performance and blocking behavior. I saw approx 10% to 15% improvement running 1080p 30FPS MP4 transcoding when buffering was added on a 4GB Pi 4B writing to ramdisk.

Project `MMALSharp.FFmpeg`:
- ffmpeg class is now only helper functions
- converted the existing AVI and RTMP helper functions
- added an MP4 helper function
- added a `VLCCaptureHandler` class with VLC MJPEG helper function
- helper functions return a configured `ExternalProcessCaptureHandler`

Project `MMALSharp.Tests`:
- updated the one existing ffmpeg test

Questions:
- should the need to await two Tasks be internalized into the handler?
- internalizing would require referencing the camera or `IOutputPort` assembly
- VLC example seems useful, but weird as part of the FFmpeg project...

Usage versus the usual one-liner call to `await ProcessAsync(...)`:

```csharp
await Task.WhenAll(new Task[]{
    ffmpeg.ManageProcessLifecycleAsync(timerToken.Token),
    cam.ProcessAsync(cam.Camera.VideoPort, timerToken.Token),
}).ConfigureAwait(false);
```

### DIFF
--- a/src/MMALSharp.FFmpeg/Handlers/FFmpegCaptureHandler.cs
+++ b/src/MMALSharp.FFmpeg/Handlers/FFmpegCaptureHandler.cs
@@ -3,34 +3,32 @@
 // Licensed under the MIT License. Please see LICENSE.txt for License info.
 // </copyright>
 
-using System;
-using System.Diagnostics;
-using System.Text;
-using MMALSharp.Common;
-
 namespace MMALSharp.Handlers
 {
     /// <summary>
-    /// Experimental FFmpeg specific capture handler.
+    /// Helper-methods for invoking ExternalProcessCaptureHandler to output to VLC.
     /// </summary>
-    public class FFmpegCaptureHandler : IVideoCaptureHandler
+    public static class FFmpegCaptureHandler
     {
-        private Process _process;
-        
-        /// <summary>
-        /// The total size of data that has been processed by this capture handler.
-        /// </summary>
-        protected int Processed { get; set; }
-                
         /// <summary>
         /// Streams video from the standard output stream via FFmpeg to an RTMP server.
         /// </summary>
         /// <param name="streamName">The meta name of the stream.</param>
         /// <param name="streamUrl">The url of your RTMP server - the url to stream to.</param>
-        /// <returns>A new instance of <see cref="FFmpegCaptureHandler"/> with process arguments to push to an RTMP stream.</returns>
-        public static FFmpegCaptureHandler RTMPStreamer(string streamName, string streamUrl)
+        /// <param name="echoOutput">Whether to echo stdout and stderr to the console or suppress it. Defaults to true.</param>
+        /// <returns>An initialized instance of <see cref="ExternalProcessCaptureHandler"/></returns>
+        public static ExternalProcessCaptureHandler RTMPStreamer(string streamName, string streamUrl, bool echoOutput = true)
         {
-            return new FFmpegCaptureHandler($"-i - -vcodec copy -an -f flv -metadata streamName={streamName} {streamUrl}");
+            var opts = new ExternalProcessCaptureHandlerOptions
+            {
+                Filename = "ffmpeg",
+                Arguments = $"-i - -vcodec copy -an -f flv -metadata streamName={streamName} {streamUrl}",
+                EchoOutput = echoOutput,
+                DrainOutputDelayMs = 500, // default
+                TerminationSignals = ExternalProcessCaptureHandlerOptions.signalsFFmpeg
+            };
+
+            return new ExternalProcessCaptureHandler(opts);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


techyian:
Many thanks for sending in this PR Jon. I'm assuming your testing so far has been done in a .NET Core environment? I'll just need to make sure everything is OK on Mono also. I'll try and find some time this week to have a look over it all.

MV10:
Yes, purely .NET Core. Hadn't thought about that angle, sorry. Haven't touched Mono since the wife and I dabbled in game dev in Unity years ago, not really set up for that.

MV10:
I see your `CancellationToken.AsTask` extension, that's what my handler is doing in the `WaitForCancellationAsync` method (albeit as a completion). I'll push a change to use your method.

src/MMALSharp.Processing/Handlers/ExternalProcessCaptureHandler.cs
```diff
@@ -0,0 +1,330 @@
...
+    /// <summary>
+    /// Options to pass to the constructor of <see cref="ExternalProcessCaptureHandler"/>.
+    /// </summary>
+    public class ExternalProcessCaptureHandlerOptions
+    {
+        /// <summary>
+        /// In theory, ffmpeg responds to a pair of SIGINT signals with a clean shutdown, although in
+        /// practice this doesn't appear to work when ffmpeg is running as a child process.
+        /// </summary>
+        public static Signum[] signalsFFmpeg = new[] { Signum.SIGINT, Signum.SIGINT };
```
techyian: Please can these static variables start with an uppercase "S"?

src/MMALSharp.Processing/Handlers/ExternalProcessCaptureHandler.cs
```diff
@@ -0,0 +1,330 @@
...
+        // result of an operation to some external "observer" -- but by definition the caller that fires
+        // an event doesn't care about the result of the operation. The only caveat is that you get no
+        // exception handling; an unhandled exception would terminate the process.
+        private async void WriteToBuffer(object sendingProcess, DataReceivedEventArgs e)
+        {
+            try
+            {
+                // Technically the faster TryWrite method is guaranteed to work for an
+                // unbounded channel, but in this case non-blocking async is more important.
+                if (_stdoutBuffer != null && e.Data != null)
```
techyian, line 229: To keep with the styling of the rest of the project, please can you make sure that any conditional block includes curly braces.

src/MMALSharp.Processing/Handlers/ExternalProcessCaptureHandler.cs
```diff
@@ -0,0 +1,330 @@
...
+        /// The total size of data that has been processed by this capture handler.
+        /// </summary>
+        protected int Processed { get; set; }
+
+        /// <summary>
+        /// Creates a new instance of <see cref="ExternalProcessCaptureHandler"/> with the specified options.
+        /// </summary>
+        public ExternalProcessCaptureHandler(ExternalProcessCaptureHandlerOptions options)
+        {
+            if(MMALLog.Logger != null)
```
techyian: You don't need to include the null check as the `MMALLog.Logger` class handles this on your behalf so you can safely remove.

src/MMALSharp.FFmpeg/Handlers/VLCCaptureHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// </summary>
+        /// <param name="listenPort">The port to listen on. Defaults to 8554.</param>
+        /// <param name="echoOutput">Whether to echo stdout and stderr to the console or suppress it. Defaults to true.</param>
+        /// <returns>An initialized instance of <see cref="ExternalProcessCaptureHandler"/></returns>
+        public static ExternalProcessCaptureHandler StreamMJPEG(int listenPort = 8554, bool echoOutput = true)
+        {
+            var opts = new ExternalProcessCaptureHandlerOptions
+            {
+                Filename = "cvlc",
+                Arguments = $"stream:///dev/stdin --sout \"#transcode{{vcodec=mjpg,vb=2500,fps=20,acodec=none}}:standard{{access=http{{mime=multipart/x-mixed-replace;boundary=7b3cc56e5f51db803f790dad720ed50a}},mux=mpjpeg,dst=:{listenPort}/}}\" :demux=h264",
```
techyian: A few comments about this:

1) I can see the bitrate and fps values are static here, do we need to have a way of retrieving them from this class? It could be that the `MMALCameraConfig` class is moved into `MMALSharp.Common`, it might be a bit cleaner than adding additional parameters to the method?
2) Can you provide any additional context around the "boundary" argument? I have looked online and can see this same GUID referenced in many places but no context as to what this GUID represents.

src/MMALSharp.FFmpeg/Handlers/FFmpegCaptureHandler.cs
```diff
@@ -3,166 +3,82 @@
...
-        /// <inheritdoc />
-        public void Dispose()
-        {
-            if (!_process.HasExited)
+            var opts = new ExternalProcessCaptureHandlerOptions
             {
-                _process.Kill();
-            }
+                Filename = "ffmpeg",
+                Arguments = $"-framerate 24 -i - -b:v 2500k -c copy -movflags +frag_keyframe+separate_moof+omit_tfhd_offset+empty_moov {directory.TrimEnd()}/{filename}.mp4",
```
techyian: Same comments around framerate and bitrate here.

src/MMALSharp.FFmpeg/Handlers/FFmpegCaptureHandler.cs
```diff
@@ -3,166 +3,82 @@
...
-using System;
-using System.Diagnostics;
-using System.Text;
-using MMALSharp.Common;
-
 namespace MMALSharp.Handlers
 {
     /// <summary>
-    /// Experimental FFmpeg specific capture handler.
+    /// Helper-methods for invoking ExternalProcessCaptureHandler to output to VLC.
```
techyian: I think this comment may have a typo, shouldn't  "VLC" be "FFmpeg"?

techyian (changes requested):
Hi Jon, 

I've had chance to look over your PR and I'm very impressed with your work, thank you very much for submitting it. There's just a few comments I've added, mostly styling related to match the rest of the project. I've done a little bit of testing on .NET Core and Mono and it seems to be working as expected on both platforms. I had noticed in my testing that when recording for 3 minutes, processing with FFmpeg doesn't actually complete for a further 30 seconds but this may be pre-existing. Once you've made the relevant changes and so long as nothing else crops up in the meantime I'm happy to get this merged in.

Thanks, Ian.

src/MMALSharp.FFmpeg/Handlers/VLCCaptureHandler.cs
MV10: 1. That bitrate and fps caps the MJPEG output stream, it isn't related to the input/camera settings. If you have a busy wifi, you could set those lower to reduce congestion. The browser will try to pull MJPEG frames back as fast as possible, it's up to the server how quickly a new frame is available to send. So probably arguments to the call make more sense.

2. The value to the boundary argument is arbitrary and is part of the MIME-type declaration in the response headers. VLC just happens to hard-code that particular string. I have wondered why, so I did some searching, but came up empty-handed.

Boundaries are discussed in this SO post and reply: https://stackoverflow.com/questions/47729941/mjpeg-over-http-specification

There's a small bug though, that GUID should be prefixed by two dashes. The answer on that SO post explains the two dashes are part of the MIME RFC, so I'll add them back. I saw the boundary marker referenced online both with and without dashes, and it works without them, so I left them off, figuring less is more. I've also seen people use varying longer strings of dashes, so I suppose it's not well understood. Anyway, I'll fix it.

src/MMALSharp.FFmpeg/Handlers/FFmpegCaptureHandler.cs
MV10: Edit -- forgot this call was ffmpeg transcoding.

This is the same situation, it defines the output of the transcoding, so I'll implement with arguments.

If we did need or want source stream values, moving `MMALCameraConfig` would be problematic, it has many dependencies from the MMALSharp project.

MV10:
I'm also fixing up a couple places where I overlooked `ConfigureAwait(false)` since you're using Mono (apparently it's no longer necessary under .NET Core, so I've been trying to wean myself off the habit), and I'm moving the handler's options class to a separate file which is a better match for the rest of the codebase.

src/MMALSharp.Processing/Handlers/ExternalProcessCaptureHandler.cs
```diff
@@ -0,0 +1,266 @@
...
+                    Syscall.kill(_process.Id, sigint);
+                }
+            }
+
+            // gives output buffer a chance to drain, and also allows the process to do exit-cleanup
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+            while (!_process.HasExited && stopwatch.ElapsedMilliseconds < _options.DrainOutputDelayMs)
+            {
+                await Task.Delay(50);
```
techyian: Thanks Jon for your updates and well spotted about `ConfigureAwait`. These `await` calls will also need it adding too please.

src/MMALSharp.Processing/Handlers/ExternalProcessCaptureHandler.cs
```diff
@@ -0,0 +1,266 @@
...
+        // exception handling; an unhandled exception would terminate the process.
+        private async void WriteToBuffer(object sendingProcess, DataReceivedEventArgs e)
+        {
+            try
+            {
+                // Technically the faster TryWrite method is guaranteed to work for an
+                // unbounded channel, but in this case non-blocking async is more important.
+                if (_stdoutBuffer != null && e.Data != null)
+                {
+                    await _stdoutBuffer.Writer.WriteAsync(e.Data);
```
techyian: And here too.

techyian (changes requested):
Hi Jon, thanks for your changes. I usually forget to add the `ConfigureAwait` calls initially but well spotted, there's just a couple of others that need adding which I've commented on. I know ASP.NET Core doesn't need `ConfigureAwait` anymore as there's no SynchronizationContext but I'm not sure whether the same applies to .NET Core as a platform. Once those are added I think we're good to merge! Thanks again.

MV10:
I've often wondered about the `ConfigureAwait` guidance. Since Blazor is their official UI direction including non-web scenarios (once the other renderers are added), I imagine this is why it's a blanket statement, even native renderers will be ASP.NET-based under the hood. Probably safest to continue adding it to library projects for the time being, though.

MV10:
I still don't like the need for this, particularly:

```csharp
await Task.WhenAll(new Task[]{
    ffmpeg.ManageProcessLifecycleAsync(timerToken.Token),
    cam.ProcessAsync(cam.Camera.VideoPort, timerToken.Token),
}).ConfigureAwait(false);
```

I suppose nobody could reasonably use this handler without RTFM but it's still a break from the pattern. I played with a custom `ProcessAsync` on the handler but that was also just weird and required a tangle of new dependencies. What if we added an optional extra Task like this:

```csharp
ProcessAsync(IOutputPort p, CancellationToken t = default, Task concurrentTask = null)
```

Then running the capture would look like this:

```csharp
await cam.ProcessAsync(cam.Camera.VideoPort, cts.Token, ffmpeg.ManageProcessLifecycleAsync(cts.Token));
```

Internally `ProcessAsync` would become a wrapper which either awaited the "real" process call, or used the `WhenAll` form if the optional extra Task is provided.

What do you think? Too much tail-wagging-the-dog? I tried the other way around -- making the handler's lifecycle call accept the camera processing Task, but that is probably even weirder from the breaking-the-pattern standpoint (e.g. you're no longer using the camera to run processing). And by adding it on the camera side as a general-purpose Task it opens it up to coordinating other concurrent processing in the future. It could then even be overloaded to accept an array of Tasks in that same arg position to process any number of tasks.

Edit -- keeping in mind that returns a hot task (already running) which is no problem for this handler. But for better synchronization a `Func<Task>` might be preferable. Gets a little ugly because of the need for an async lambda to pass args though. I suppose with overloads we could add both -- and array variants.

```csharp
await cam.ProcessAsync(cam.Camera.VideoPort, cts.Token, async () => ffmpeg.ManageProcessLifecycleAsync(cts.Token));
```

MV10:
I'm looking through `ProcessAsync` today and seeing that all your components are already managed as concurrent tasks. What may be easier (and better-hide this handler's requirement from the library consumer) is a new interface which is used to provide `ProcessAsync` with one or more arbitrary additional tasks that the handler requires.

I think it could be as simple as `ProcessAsync` tries to cast each component to something like `IProcessHandlerTasks` (or whatever) and if it succeeds, it passes that internal `List<Task>` to an interface method like `AddTasks` before processing starts.

Clean, simple, 100% transparent to the library consumer. What do you think?

techyian:
I hear what you're saying, but what you're trying to achieve here doesn't really affect the components themselves, it affects the capture handlers. Each port has a "Callback Handler" associated with it to provide a layer of abstraction, which in turn has a capture handler association. I'm not sure it'll be as simple as you're suggesting.

The `ProcessAsync` method really has one job and that is to handle the processing of data that is returned by the camera (or standalone data) and asynchronously wait for the `TaskCompletionSource` objects to trigger. I see the ExternalCaptureHandler's task that we're wanting to await as something totally separate to the running of MMALSharp and although modifying `ProcessAsync` and any other related code to support your suggestions would keep things cleaner for the end-user, it straps on additional responsibility to this method and might make things harder to maintain going forward.

FFmpeg support has always been experimental and was added primarily as a shortcut for users so they didn't have to manually run it as a separate process once MMALSharp had finished doing its thing. It is a separate application and I think it's totally acceptable to expect users to read the manual before using this functionality - the way you have designed it actually makes it clearer in my head that these are two separate tasks complimenting each other. Additional code comments can be added if you feel they'll be necessary or changing method names (for example, does `ProcessExternalAsync` work better than `ManageProcessLifecycleAsync`?) but I'm quite happy with this approach as it currently stands.

MV10:
Sounds good to me. Thanks for the clarification.

techyian:
I've just edited my post, sorry if you missed it before replying. I was just wondering whether using the method name `ProcessExternalAsync` may be better than `ManageProcessLifecycleAsync`, what do you think to that or do you think it's better as it currently stands?

MV10:
I did miss that, I'll change it. Funny too, I actually used `ProcessExternalAsync` for a short time, but originally had the "outer" code doing more of the work and changed it trying to clarify what was going on. Now that it's back to just one extra step, I think that name is good, I feel like it sort of helps "relate" it to the camera `ProcessAsync` call.

techyian (approved):
Thanks Jon. Let's get this merged.

PR #165, opened 2020-08-15
On-demand images, take two

### LINKED ISSUE #163: Write JPGs during motion-detection recording
For the past couple of days, I've been trying to alter the motion detection example to also capture JPEG images periodically. My use-case is to match my existing CCTV DVR, which captures one image every second for the first 3 seconds after motion-detection begins. These are emailed and/or messaged to us.

The wiki shows how to "manually" take a snapshot, but that relies upon calling `ProcessAsync` for the camera's `StillPort`, which is obviously not compatible with motion detection. The wiki also shows how to continuously write both video and stills (the `VideoAndImages` example), but there doesn't appear to be a way to control when a still frame is captured while recording. Actually, I couldn't get continuous still-capture working with the motion detection sample either, it always wrote a single zero-length JPEG (I assume when `ProcessAsync` gets everything started). When I try to wire it up to the splitter similar to the video handler/encoder, I get MMAL errors (I think because the splitter output is pointed to the camera video port).

Failed attempts aside, is this combination possible with the current image encoder API? If not, could I perhaps modify `MMALImageEncoder` to write a JPG on demand? I haven't taken a look at the encoder for myself yet, as I clearly don't even understand how to get it working in the splitter pipeline -- which is probably another question I should ask: is the existing pipeline API compatible with this combination?

MV10:
This is sort of a cross-pollination of my previous effort and the existing capture handlers. As mentioned in the issue, it gives us the on-demand single-image capture capability I wanted, it retains the continuous-capture capability of the existing handler, and requires no extra classes or interfaces. It should also perform very marginally better, it shouldn't accidentally store incomplete frames, and the timestamp format is configurable when writing those types of filenames. 

`IFileStreamCaptureHandler`
* Adds the `NewFrame` method to be invoked when the callback reaches EOS

`FastImageOutputCallbackHandler`
* At EOS, invokes `NewFrame` instead of `NewFile`

`ImageStreamCaptureHandler`
* Adds optional `continuousCapture` argument to constructors to pass to the base

`FileStreamCaptureHandler`
* Derives from `MemoryStreamCaptureHandler`
* `CaptureNextFrame` bool property, resets after one write if `CaptureNextFrame` is false
* `ContinuousCapture` bool property, when true also keeps `CaptureNextFrame` true
* `FilenameDateTimeFormat` string property
* `LastWrittenPathname` string property
* `continuousCapture = true` argument to constructors to keep existing behavior
* Use last pathname for `GetFilename` and `GetFilepath`
* `NewFile` writes on demand and only when invoked by `NewFrame` (which is automatic)
* `NewFrame` manages the capture properties and resets the `MemoryStream`
* Private field `_skippingFirstPartialFrame` bool, true until `NewFrame` is invoked once

Conceptually, the only big change is that no "ambient" file stream is open, so it doesn't make sense to query for the "current" filename/path. Performance should be identical (or better) since file streams are just memory streams under the hood with some occasional flush behaviors. This also means `CurrentFilename` is only set when the filepath-based constructor is used.

Logically, `NewFrame` (invoked at EOS) checks if a capture is requested. If so, `NewFile` is invoked. If continuous capture is disabled, the capture flag is also disabled. Meanwhile, if `NewFile` is invoked externally, but next-frame capture is not enabled, the flag is set to true which waits for EOS, then `NewFrame` invokes `NewFile`.

Fixes #163

### DIFF
--- a/src/MMALSharp.Processing/Handlers/FileStreamCaptureHandler.cs
+++ b/src/MMALSharp.Processing/Handlers/FileStreamCaptureHandler.cs
@@ -9,16 +9,47 @@ using System;
 using System.Collections.Generic;
 using System.IO;
 using System.Linq;
+using System.Runtime.CompilerServices;
 
 namespace MMALSharp.Handlers
 {
     /// <summary>
     /// Processes image data to a <see cref="FileStream"/>.
     /// </summary>
-    public class FileStreamCaptureHandler : StreamCaptureHandler<FileStream>, IFileStreamCaptureHandler
+    public class FileStreamCaptureHandler : MemoryStreamCaptureHandler, IFileStreamCaptureHandler
     {
         private readonly bool _customFilename;
         private int _increment;
+        private bool _skippingFirstPartialFrame = true;
+        private bool _continuousCapture;
+
+        /// <summary>
+        /// When true, the next full frame will be written. If <see cref="ContinuousCapture"/> is not also
+        /// true, this property will be reset to false after writing the image so that only one image is written.
+        /// </summary>
+        public bool CaptureNextFrame { get; set; }
+
+        /// <summary>
+        /// When true, every frame is written to storage.
+        /// </summary>
+        public bool ContinuousCapture 
+        {
+            get => _continuousCapture; 
+
+            set
+            {
+                _continuousCapture = value;
+                if(_continuousCapture)
+                {
+                    CaptureNextFrame = true;
+                }
+            }
+        }
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


MV10:
I was just thinking about that comment in your code about not calling `File.Exists` repeatedly. I think that could be resolved by comparing the new filename to the most-recently-written filename. If they're the same, then you start adding the auto-increment counter. I think that way even very high-speed writes (like 4X binning) would have relatively high throughput. We may not need the other pre-defined filename approach at all (though I can see how it would be desirable for other reasons).

PR #171, opened 2020-08-20
Improved motion perf

### LINKED ISSUE #170: Additional motion detection enhancements
Hey Ian, some interesting news... back in #161 the motion detection processing improved a lot -- my test setup was getting 25ms per pass, about 255 frames over a 30 second run. I've managed to more than double that ... about 10.5ms per pass, and more than 700 frames over a 30 second run.

I'm actually setting this up as a prelude to attempt fine-grained-cell proximity motion detection (e.g. "ignore my small dog" mode), but these changes are sufficiently substantial to warrant a PR of their own.

The crazy part is that the changes are simple. The largest change is to simply pre-calculate the `Rectangle` objects once along with a few other one-time optimizations that the existing 0.7 build does repeatedly on the fly. Apparently even struct allocs can be expensive. Nobody really needs 10ms motion detection (probably), so my hope is that this headspace will allow for more complex analysis. Also I want to investigate the use of `Parallel.For` versus arrays of tasks -- in .NET Core they made good progress generalizing task parallelism but `Parallel.For` is still theoretically optimal for CPU-bound operations like these.

It's quite late and I need to turn in, but I have a few other things in mind before I PR this. But the improvement was so substantial I thought I'd open a new discussion as a heads up.

```csharp
public class ProximityDiffAnalyser : FrameAnalyser
{
    // When true, PrepareTestFrame does additional start-up processing
    private bool _firstFrame = true;

    // Frame dimensions collected when the first full frame is complete
    private int _frameWidth;
    private int _frameHeight;
    private int _frameStride;
    private int _frameBpp;

    private byte[] _mask;
    private Stopwatch _testFrameAge;

    // new stuff for dev/test
    public int _cellDivisor = 2;
    private int _cellCount;
    private int _cellWidth;
    private int _cellHeight;
    private Rectangle[] _cellRect;
    private Task[] _diffTask;

    // performance testing
    public long ElapsedTime;
    public long Passes;

    internal Action OnDetect { get; set; }

    /// <summary>
    ///  This is the image we are comparing against new incoming frames.
    /// </summary>
    protected byte[] TestFrame { get; set; }

    /// <summary>
    /// Indicates whether we have a full test frame.
    /// </summary>
    protected bool FullTestFrame { get; set; }

    /// <summary>
    /// The motion configuration object.
    /// </summary>
    protected MotionConfig MotionConfig { get; set; }

    /// <summary>
    /// The image metadata.
    /// </summary>
    protected ImageContext ImageContext { get; set; }

    /// <summary>
    /// Creates a new instance of <see cref="ProximityDiffAnalyser"/>.
    /// </summary>
    /// <param name="config">The motion configuration object.</param>
    /// <param name="onDetect">A callback when changes are detected.</param>
    public ProximityDiffAnalyser(MotionConfig config, Action onDetect)
    {
        this.MotionConfig = config;
        this.OnDetect = onDetect;

        _testFrameAge = new Stopwatch();
    }

    /// <inheritdoc />
    public override void Apply(ImageContext context)
    {
        this.ImageContext = context;

        base.Apply(context);

        if (!this.FullTestFrame)
        {
            if (context.Eos)
            {
                this.FullTestFrame = true;
                this.PrepareTestFrame();
                MMALLog.Logger.LogDebug("EOS reached for test frame.");
            }
        }
        else
        {
            MMALLog.Logger.LogDebug("Have full test frame.");

            if (this.FullFrame && !this.TestFrameExpired())
            {
                MMALLog.Logger.LogDebug("Have full frame, checking for changes.");

                this.CheckForChanges(this.OnDetect);
            }
        }
    }

    /// <summary>
    /// Resets the test and working frames this analyser is using.
    /// </summary>
    public void ResetAnalyser()
    {
        this.TestFrame = null;
        this.WorkingData = new List<byte>();
        this.FullFrame = false;
        this.FullTestFrame = false;

        _testFrameAge.Reset();
    }

    private void PrepareTestFrame()
    {
        if (_firstFrame)
        {
            // one-time collection of basic frame dimensions
            _frameWidth = this.ImageContext.Resolution.Width;
            _frameHeight = this.ImageContext.Resolution.Height;
            _frameBpp = this.GetBpp() / 8;
            _frameStride = this.ImageContext.Stride;

            // one-time setup of the diff cell parameters and arrays
            _cellCount = (int)Math.Pow(_cellDivisor, 2);
            _cellWidth = _frameWidth / _cellDivisor;
            _cellHeight = _frameHeight / _cellDivisor;
            _cellRect = new Rectangle[_cellCount];
            _diffTask = new Task[_cellCount];

            int i = 0;
            for (int row = 0; row < _cellDivisor; row++)
            {
                int y = row * _cellHeight;
                for (int col = 0; col < _cellDivisor; col++)
                {
                    int x = col * _cellWidth;
                    _cellRect[i] = new Rectangle(x, y, _cellWidth, _cellHeight);
                    i++;
                }
            }

            this.TestFrame = this.WorkingData.ToArray();

            if (!string.IsNullOrWhiteSpace(this.MotionConfig.MotionMaskPathname))
            {
                this.PrepareMask();
            }

            _firstFrame = false;
        }
        else
        {
            this.TestFrame = this.WorkingData.ToArray();
        }

        if (this.MotionConfig.TestFrameInterval != TimeSpan.Zero)
        {
            _testFrameAge.Restart();
        }
    }

    private int GetBpp()
    {
        PixelFormat format = default;

        // RGB16 doesn't appear to be supported by GDI?
        if (this.ImageContext.PixelFormat == MMALEncoding.RGB24)
        {
            return 24;
        }

        if (this.ImageContext.PixelFormat == MMALEncoding.RGB32 || this.ImageContext.PixelFormat == MMALEncoding.RGBA)
        {
            return 32;
        }

        if (format == default)
        {
            throw new Exception("Unsupported pixel format.");
        }

        return 0;
    }

    private void PrepareMask()
    {
        using (var fs = new FileStream(this.MotionConfig.MotionMaskPathname, FileMode.Open, FileAccess.Read))
        using (var mask = new Bitmap(fs))
        {
            // Verify it matches our frame dimensions
            var maskBpp = Image.GetPixelFormatSize(mask.PixelFormat) / 8;
            if (mask.Width != _frameWidth || mask.Height != _frameHeight || maskBpp != _frameBpp)
            {
                throw new Exception("Motion-detection mask must match raw stream width, height, and format (bits per pixel)");
            }

            // Store the byte array
            BitmapData bmpData = null;
            try
            {
                bmpData = mask.LockBits(new Rectangle(0, 0, mask.Width, mask.Height), ImageLockMode.ReadOnly, mask.PixelFormat);
                var pNative = bmpData.Scan0;
                int size = bmpData.Stride * mask.Height;
                _mask = new byte[size];
                Marshal.Copy(pNative, _mask, 0, size);
            }
            finally
            {
                mask.UnlockBits(bmpData);
            }
        }
    }

    private bool TestFrameExpired()
    {
        if (this.MotionConfig.TestFrameInterval == TimeSpan.Zero || _testFrameAge.Elapsed < this.MotionConfig.TestFrameInterval)
        {
            return false;
        }

        MMALLog.Logger.LogDebug("Have full frame, updating test frame.");
        this.PrepareTestFrame();
        return true;
    }

    private void CheckForChanges(Action onDetect)
    {
        Passes++;
        var sw = new Stopwatch();
        sw.Start();

        var diff = this.Analyse();

        sw.Stop();
        ElapsedTime += sw.ElapsedMilliseconds;

        if (diff >= this.MotionConfig.Threshold)
        {
            MMALLog.Logger.LogInformation($"Motion detected! Frame difference {diff}.");
            onDetect();
        }
    }

    private int Analyse()
    {
        var currentBytes = this.WorkingData.ToArray();

        int diff = 0;

        for (int i = 0; i < _cellCount; i++)
        {
            var capture = i;
            _diffTask[i] = Task.Run(() =>
            {
                diff += this.CheckDiff(_cellRect[capture], currentBytes);
            });
        }

        Task.WaitAll(_diffTask);

        return diff;
    }

    private int CheckDiff(Rectangle cell, byte[] currentFrame)
    {
        int diff = 0;

        for (int col = cell.X; col < cell.X + cell.Width; col++)
        {
            for (int row = cell.Y; row < cell.Y + cell.Height; row++)
            {
                var index = (col * _frameBpp) + (row * _frameStride);

                if (_mask != null)
                {
                    var rgbMask = _mask[index] + _mask[index + 1] + _mask[index + 2];

                    if (rgbMask == 0)
                    {
                        continue;
                    }
                }

                var rgb1 = TestFrame[index] + TestFrame[index + 1] + TestFrame[index + 2];

                var rgb2 = currentFrame[index] + currentFrame[index + 1] + currentFrame[index + 2];

                if (rgb2 - rgb1 > MotionConfig.Threshold)
                {
                    diff++;
                }

                // If the threshold has been exceeded, we want to exit from this method immediately for performance reasons.
                if (diff > MotionConfig.Threshold)
                {
                    return diff;
                }
            }

            if (diff > MotionConfig.Threshold)
            {
                return diff;
            }
        }

        return diff;
    }
}
```

MV10:
Motion-detection performance improvements, fixes #170

### DIFF
--- a/src/MMALSharp.Processing/Processors/Motion/FrameDiffAnalyser.cs
+++ b/src/MMALSharp.Processing/Processors/Motion/FrameDiffAnalyser.cs
@@ -43,7 +43,7 @@ namespace MMALSharp.Processors.Motion
         /// value (so the default of 8 yields 64 cells). These cells are processed in parallel. This
         /// should be a value that divides evenly into the X and Y resolutions of the motion stream.
         /// </summary>
-        public int CellDivisor { get; set; } = 8;
+        public int CellDivisor { get; set; } = 32;
 
         internal Action OnDetect { get; set; }
 
--- a/src/MMALSharp.Processing/Processors/Motion/FrameDiffAnalyser.cs
+++ b/src/MMALSharp.Processing/Processors/Motion/FrameDiffAnalyser.cs
@@ -34,13 +34,8 @@ namespace MMALSharp.Processors.Motion
         private byte[] _mask;
         private Stopwatch _testFrameAge;
 
-        private struct DiffRect
-        {
-            internal int diff;
-            internal Rectangle rect;
-        }
-
-        private DiffRect[] _cell;
+        private int[] _cellDiff;
+        private Rectangle[] _cellRect;
         private byte[] _workingData;
 
         /// <summary>
@@ -138,7 +133,9 @@ namespace MMALSharp.Processors.Motion
                 _frameStride = this.ImageContext.Stride;
 
                 // one-time setup of the diff cell parameters and arrays
-                _cell = new DiffRect[(int)Math.Pow(CellDivisor, 2)];
+                int indices = (int)Math.Pow(CellDivisor, 2);
+                _cellRect = new Rectangle[indices];
+                _cellDiff = new int[indices];
                 int cellWidth = _frameWidth / CellDivisor;
                 int cellHeight = _frameHeight / CellDivisor;
                 int i = 0;
@@ -148,7 +145,7 @@ namespace MMALSharp.Processors.Motion
                     for (int col = 0; col < CellDivisor; col++)
                     {
                         int x = col * cellWidth;
-                        _cell[i].rect = new Rectangle(x, y, cellWidth, cellHeight);
+                        _cellRect[i] = new Rectangle(x, y, cellWidth, cellHeight);
                         i++;
                     }
                 }
@@ -252,7 +249,7 @@ namespace MMALSharp.Processors.Motion
         {
             _workingData = this.WorkingData.ToArray();
 
-            var result = Parallel.ForEach(_cell, (src, loopState) => CheckDiff(src, loopState));
+            var result = Parallel.ForEach(_cellDiff, (cell, loopState, loopIndex) => CheckDiff(loopIndex, loopState));
 
             // How Parallel Stop works: https://docs.microsoft.com/en-us/previous-versions/msp-n-p/ff963552(v=pandp.10)#parallel-stop
             if (!result.IsCompleted && !result.LowestBreakIteration.HasValue)
@@ -262,16 +259,16 @@ namespace MMALSharp.Processors.Motion
             else
             {
                 int diff = 0;
-                foreach (var cell in _cell)
-                    diff += cell.diff;
+                foreach (var cellDiff in _cellDiff)
+                    diff += cellDiff;
                 return diff;
             }
         }
 
-        private void CheckDiff(DiffRect cell, ParallelLoopState loopState)
+        private void CheckDiff(long cellIndex, ParallelLoopState loopState)
         {
-            cell.diff = 0;
-            var rect = cell.rect;
+            int diff = 0;
+            var rect = _cellRect[cellIndex];
 
             for (int col = rect.X; col < rect.X + rect.Width; col++)
             {
@@ -295,23 +292,27 @@ namespace MMALSharp.Processors.Motion
 
                     if (rgb2 - rgb1 > MotionConfig.Threshold)
                     {
-                        cell.diff++;
+                        diff++;
                     }
 
                     // If the threshold has been exceeded, exit immediately and preempt any CheckDiff calls not yet started.
-                    if (cell.diff > MotionConfig.Threshold)
+                    if (diff > MotionConfig.Threshold)
                     {
+                        _cellDiff[cellIndex] = diff;
                         loopState.Stop();
                         return;
                     }
                 }
 
-                if (cell.diff > MotionConfig.Threshold)
+                if (diff > MotionConfig.Threshold)
                 {
+                    _cellDiff[cellIndex] = diff;
                     loopState.Stop();
                     return;
                 }
             }
+
+            _cellDiff[cellIndex] = diff;
         }
     }
 }
--- a/src/MMALSharp.Processing/Processors/Motion/FrameDiffAnalyser.cs
+++ b/src/MMALSharp.Processing/Processors/Motion/FrameDiffAnalyser.cs
@@ -34,6 +34,22 @@ namespace MMALSharp.Processors.Motion
         private byte[] _mask;
         private Stopwatch _testFrameAge;
 
+        private struct DiffRect
+        {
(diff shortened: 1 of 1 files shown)


MV10:
Question for you, Ian:

Currently it aborts processing as soon as a single cell diff exceeds the threshold, which makes sense. But I want to "post-process" the cell diffs, which means I want all cells processed. Would you be open to a flag of some kind to disable early exit so that I can subclass for post-processing? I'd also have to change the access of cell data fields and `CheckForChanges` which are currently `private`.

I'm thinking `protected` rather than `public`, although maybe a motion detection consumer might have a use for the cell data in response to the `onDetect` callback? But that's probably an edge case, and for performance reasons we'd want to keep them as fields rather than properties, which argues for `protected` access.

I should clarify I'm proposing this for some future PR, if I can work out proximity processing in some useful way -- unless you think `public` is worthwhile in which case I'll do it now.

Also, if you think this is worthwhile, are we getting into territory where a diff-specific interface might be sensible?

techyian:
> Would you be open to a flag of some kind to disable early exit so that I can subclass for post-processing?

I don't have any issues with this, I'd suggest adding it to the `MotionConfig` class. I would prefer protected access rather than opening everything up. Are you proposing making any of the methods `virtual` or would you do your post processing in the `Apply` method? 

> Also, if you think this is worthwhile, are we getting into territory where a diff-specific interface might be sensible?

Potentially, yes. At this stage I'm happy with how this PR is looking, I'm really impressed at how low you've got the latency so I'm going to get this merged into dev now. Due to the change in #169 where the `FrameBufferCaptureHandler` is now doing the motion detection, it might be worthwhile adding the interface in that branch and making the relevant changes required to call `Apply` on a `IFrameDiffAnalyser` interface (or other suitable name).

MV10:
Sounds good. I was thinking of making `CheckForChanges` virtual and part of the diff interface.

I think we're home all day tomorrow, I'll work on it then.

MV10:
> I'm going to get this merged into dev now

If you would do the honors, sir... (Since I now have two branches impacting `MMALSharp.Processing` I can't easily continue the discussed changes in the other PR until this is merged... 😄)

Issue #108, opened 2020-01-01
Image processing performance


techyian:
Image processing speed on full size frames is relatively slow. My current solution has been to split the image into 4 rectangles and run a Task against each of them to perform the processing. It works well for small images and I noticed a significant speed improvement. Unsure as to how much more I can split the image up before no further improvements are seen.

MV10:
New Year's Day? You're as bad as I am. 😀 

Does this refer to image processing such as the matrix operations, and could the `Parallel` changes made in the motion detection performance PR #171 also apply here?

techyian:
:D I don't really follow the hype of New Year, I prefer Christmas to be honest. But yes, I'd like to replicate the performance work you've done for motion detection on the matrix operation code. Could there be a common place for this code? I'd rather not duplicate it in two places.

MV10:
Starting another pass on PR #175 to get `FrameAnalyser` ready for this. I think it's going to work well.

MV10:
Ian, in parts of the motion detection code we discussed the use of fields instead of properties for speed. I don't think I mentioned the specific numbers, but if you recall the tests where it completed motion detection in 6ms per frame, I accidentally (force of habit) set up a new value as a property and that alone slowed it to 12ms per frame. Changing it back to a field brought it back to 6ms. Property accessor method calls are relatively expensive in tight, performance-sensitive loops.

It occurs to me that MMALSharp's hot-path pattern is generally `Apply(ImageContext context)` in various interfaces, and I suspect performance could be improved throughout by simply changing `ImageContext` properties into fields. I certainly understand the Computer Science 101 theory of preferring properties over fields for public members, but in this library and on these devices, I also feel that breaking the rules might be warranted if it produces significant performance improvements. (And after all, that rule of thumb in C# is mostly based on the fact that interfaces can't declare fields, and `ImageContext` isn't represented anywhere by a separate interface definition.)

Not really the point of this thread, but it's still image processing performance, and I didn't think it necessarily warranted a separate discussion. Of course, benchmarking is needed if you're open to considering this. It's such a simple change, I'll probably start wondering about it to the point I just try it to see what happens. :)

MV10:
[...]

PR #175, opened 2020-09-01
Motion detection plug-in model and analysis

MV10:
And here we go again!

### Improved summed RGB diff logic
* Use absolute value in the comparison (bugfix)
* Separate RGB diff threshold from frame-wide change-count threshold
* Switch to cell-based frame-level change counts
* Per-cell threshold as a percentage of pixels in the cell

### New Stuff
* Separate motion detection from motion detection frame buffer handling
* A plug-in approach to the motion detection algorithm
* Optional real time video analysis output from motion algorithms
* Ability to leave motion detection running while suppressing the detection callback
* Improved test frame update logic ("cool down" avoids updating during motion)

### Capture Handlers
* `IMotionCaptureHandler.DisableMotionCapture` adds a new argument `onlyDisableCallback`
* `FrameBufferCaptureHandler` implements new argument and has renamed `FrameDiffDriver`
* `MotionAnalysisCaptureHandler` writes full raw frames to a file stream, motion algorithm callback-based

### Frame Buffer Handling
* `FrameDiffDriver` handles frame buffers and invoking the selected motion detection algorithm
* `FrameDiffMetrics` is a thread safe struct used to pass parameters to the motion algorithm
* `IMotionAlgorithm` defines how the frame driver interacts with the algorithms
* `MotionAlgorithmBase` provides a couple of utility methods
* `MotionAlgorithmRGBDiff` is the parallel processing motion algorithm and analysis code
* `FrameDiffAnalyser` is gone (replaced by `FrameDiffDriver` and `IMotionAlgorithm`)

### Configuration
* The selected algorithm is referenced by the `MotionConfig.MotionAlgorithm` property
* Pass algorithm-specific parameters to the algorithm constructor
* `Threshold` is removed from `MotionConfig`

### Example
```csharp
using var motionCaptureHandler = new FrameBufferCaptureHandler();
using var resizer = new MMALIspComponent();

resizer.ConfigureInputPort(new MMALPortConfig(MMALEncoding.OPAQUE, MMALEncoding.I420), cam.Camera.VideoPort, null);
resizer.ConfigureOutputPort<VideoPort>(0, new MMALPortConfig(MMALEncoding.RGB24, MMALEncoding.RGB24, width: 640, height: 480), motionCaptureHandler);
cam.Camera.VideoPort.ConnectTo(resizer);

await Task.Delay(2000); // camera warmup

var cts = new CancellationTokenSource(TimeSpan.FromSeconds(totalSeconds));

// These are the defaults
var motionAlgorithm = new MotionAlgorithmRGBDiff(
        rgbThreshold: 200,
        cellPixelPercentage: 50,
        cellCountThreshold: 20
    );

// Optional, outputs modified frames (other algorithms could do other things?)
// For RGBDiff algorithm, output goes nowhere without the callback
// MotionAnalysisCaptureHandler does this internally, writes output to stream
motionAlgorithm.EnableAnalysis(analysisFramebufferCallback: null);

// These are the defaults
var motionConfig = new MotionConfig(
        algorithm: motionAlgorithm, 
        testFrameInterval: TimeSpan.FromSeconds(3), 
        testFrameCooldown: TimeSpan.FromSeconds(3)
    );

    motionCaptureHandler,
    motionConfig,
    async () =>
    {
        // callback code omitted... 
    }
```


MV10:
A small but unrelated change that I'd like to commit, if you don't mind, is to activate the "generate package on build" option on the csproj files. This makes it much easier to work with the library in my separate test program.

I'm also interested in how you address that problem yourself. In my case, I have a `dev_packages` folder as a NuGet package source, and I've simply written a batch file that copies the nupkg files there and removes the packages from the user-directory .nuget cache. It's a bit inelegant though, because I have to remember to run the batch and also to close VS and re-open to get it to re-load the updated packages -- unloading and reloading the solution doesn't seem to be enough.

MV10:
Just thinking out loud (I know you're busy right now):

I am going to modify `IMotionAlgorithm` to use an output handler as the callback. The only oddity is that `ImageContext` is typically buffer-based throughout the library, whereas this will be a full raw frame (though low-res buffers are often full frame, of course). I need to think about how to populate some of the context properties (probably via `FrameDiffMetrics`).

```csharp
void EnableAnalysis(IOutputCaptureHandler handler = null);
```

While I _really_ wanted to run the analysis output (and other filtering) earlier in the pipeline so that the hardware could do h.264 against the filtered output, that isn't looking realistic. But this change still allows for streaming by piping the raw frames through `ExternalProcessCaptureHandler` and the helper in `VLCCaptureHandler`.

The new (in this PR) class `MotionAnalysisCaptureHandler` becomes unnecessary.

I'm also wondering if dishing out full frames to ffmpeg will have any effect on its ability to output a valid MP4. Probably not, but I'll give it a shot.

MV10:
Figures... according to the muxer matrix, VLC won't convert raw frames to MJPEG.

https://www.videolan.org/streaming-features.html

So ... into the ffmpeg switch-swamp I go. Wish me luck.

Annnnd... ffmpeg doesn't work as a server like VLC does. 🙄 

Getting pretty close, I had the idea of using ffmpeg to encode the RGB24 to h.264, then pipe that into VLC for streaming. After many hours of work I have it close to working. I put the commands into a script with this beast of a command line:

```
#!/bin/bash
ffmpeg -hide_banner -f rawvideo -c:v rawvideo -pix_fmt rgb24 -s:v 640x480 -r 24 -i - -f h264 -c:v h264_omx -b:v 2500k - | cvlc stream:///dev/stdin --sout "#transcode{vcodec=mjpg,vb=2500,fps=24,acodec=none}:standard{access=http{mime=multipart/x-mixed-replace;boundary=7b3cc56e5f51db803f790dad720ed50a},mux=mpjpeg,dst=:8554/}" :demux=h264
```

From a terminal I can do something like this to create a stream which I can browse to:

```
cat /media/ramdisk/input.raw | ./ffmpeg2clvc.sh
```

When I plug the script into `ExternalProcessCaptureHandler` I can see that it launches but it seems that it doesn't receive data. I suspect this is because my ffmpeg command is using the h264_omx hardware encoder but my own MMALSharp program has that tied up already...

MV10:
I overlooked your reply to #108 last week (image processing performance), but I'm starting to think `FrameAnalyser` would be a more sensible place for some of these changes (collecting frame metrics, storing the list of rectangles for parallel cell-based processing, etc.) This way the motion-diff class would only be managing the test frame, the new comparison frame, and the diff algorithm. Whereas `FrameAnalyser` would be reusable by the static image processing, or perhaps even other places in the pipeline.

I've also read that Gaussian blur is helpful for motion detection noise-removal, so once I feel this PR is ready to roll, I will probably tackle that next.

MV10:
Got it working ... this is a frame from my browser of realtime 24FPS filtered motion analysis (it's pointing elsewhere in my office and picking up the news on TV). As described earlier, /bin/bash is the external process, the raw RGB24 stream pipes into ffmpeg which turns it into an h.420 YUV420p stream (that format is what I was missing the other day) and pipes that to CLVC which demuxes into an MJPEG HTTP stream and listens on the port. Pretty cool!

image

```csharp
static async Task visualizeStream(int seconds)
{
    var cam = GetConfiguredCamera();

    MMALCameraConfig.Resolution = new Resolution(640, 480);
    MMALCameraConfig.SensorMode = MMALSensorMode.Mode7; // for some reason mode 6 has a pinkish tinge
    MMALCameraConfig.Framerate = new MMAL_RATIONAL_T(20, 1);

    Console.WriteLine("Preparing pipeline...");

    var motionAlgorithm = new MotionAlgorithmRGBDiff(
            rgbThreshold: 200,          // default = 200
            cellPixelPercentage: 50,    // default = 50
            cellCountThreshold: 20      // default = 20
        );

    var motionConfig = new MotionConfig(
            algorithm: motionAlgorithm,
            testFrameInterval: TimeSpan.FromSeconds(3), // default = 3
            testFrameCooldown: TimeSpan.FromSeconds(3)  // default = 3
        );

    var raw_to_mjpeg_stream = new ExternalProcessCaptureHandlerOptions
    {
        Filename = "/bin/bash",
        EchoOutput = true,
        Arguments = "-c \"ffmpeg -hide_banner -f rawvideo -c:v rawvideo -pix_fmt rgb24 -s:v 640x480 -r 24 -i - -f h264 -c:v libx264 -preset ultrafast -tune zerolatency -vf format=yuv420p - | cvlc stream:///dev/stdin --sout '#transcode{vcodec=mjpg,vb=2500,fps=20,acodec=none}:standard{access=http{mime=multipart/x-mixed-replace;boundary=7b3cc56e5f51db803f790dad720ed50a},mux=mpjpeg,dst=:8554/}' :demux=h264\"",
        DrainOutputDelayMs = 500, // default = 500
        TerminationSignals = ExternalProcessCaptureHandlerOptions.SignalsFFmpeg
    };

    using (var shell = new ExternalProcessCaptureHandler(raw_to_mjpeg_stream))
    using (var motion = new FrameBufferCaptureHandler(motionConfig, null))
    {
        // Argument is a capture handler, this is the bash/ffmpeg/clvc pipeline
        motionAlgorithm.EnableAnalysis(shell);

        // Not taking action when motion is detected in this example,
        // but the resizer feeds normal motion detection
        resizer.ConfigureOutputPort<VideoPort>(0, new MMALPortConfig(MMALEncoding.RGB24, MMALEncoding.RGB24, width: 640, height: 480), motion);
        cam.Camera.VideoPort.ConnectTo(resizer);

        await cameraWarmupDelay(cam);

        Console.WriteLine($"Streaming MJPEG with motion detection analysis for {seconds} sec to:");
        Console.WriteLine($"http://{Environment.MachineName}.local:8554/");
        var timeout = new CancellationTokenSource(TimeSpan.FromSeconds(seconds));
        await Task.WhenAll(new Task[]{
                shell.ProcessExternalAsync(timeout.Token),
                cam.ProcessAsync(cam.Camera.VideoPort, timeout.Token),
            }).ConfigureAwait(false);
    }

    cam.Cleanup();

    Console.WriteLine("Exiting.");
}
```

MV10:
AppVeyor can't create the ffmpeg nuget package? Looks like it skipped building the ffmpeg project, for some reason.

techyian:
This looks to be because you've added `GeneratePackageOnBuild` to true, please see [here](https://github.com/dotnet/sdk/issues/10335). To be honest, I'd rather this be removed. Locally I just reference the DLLs in my test project if there's stuff that hasn't been pushed to GitHub. The AppVeyor setup has worked well for a few years now and you can always add the MyGet repository for work that's been committed to the dev branch.

MV10:
Oops, meant to pull those out before check-in ... fixing it.

MV10:
I don't think I listed the specific changes in that last commit:
* `MotionAnalysisCaptureHandler` removed, no longer required
* `MotionAlgorithmRGBDiff` altered to output analysis-mode images to a capture handler
* Added `StreamRawRGB24asMJPEG` to the VLC helper
* Renamed the existing MJPEG streaming helper to `StreamH264asMJPEG`

If you agree with my work described in #108 and we figure out the raw capture / manipulate situation, these are the changes versus this original PR's description that I will commit. I wanted to get them written down while it's still reasonably fresh in my mind, I suspect I'm about to get swamped by my day job, too. 

### Frame and Motion Processing
* `FrameAnalyser` no longer abstract, handles first-frame metadata collection, owns parallel processing cells
* `FrameDiffMetrics` struct renamed to `FrameAnalysisMetadata` (thought it was more clear that way)
* The cell divisor was separated into horizontal and vertical counts, as explained in #108 
* `FrameDiffDriver` scaled back to just managing the mask, test frame, and invoking analysis

### Matrix Convolution
* `ConvolutionBase` requires raw frames, uses `FrameAnalyser`, uses cell-based parallel processing
* `ConvolutionBase` uses an `ImageContext` as input
* If the context `StoreFormat` property is null (via `Manipulate`), the convolution output is a raw frame
* All related classes have a constructor override to accept custom cell-count values
* `BoxBlur` processor added
* `LineDetection` processor added (directional edge detection)
* Minor `EdgeDetection` changes to make it more like the other processing classes (fewer public members)

### MMALSharp.Common
* `ImageContext` properties changed to fields for hot-path performance reasons
* `FormatRawImageExtension` adds `ImageContext.FormatRawImage` for raw `Data` to JPEG, BMP, etc

None of the above changes usage.

Again, just wanted to document everything that changed, no rush at all, having this PR open isn't blocking me or anything. I know the other thread is long and this is a lot to review! 

Have a great weekend, we're taking the dog to the bar and doing absolutely nothing useful at all! 😁

src/MMALSharp.Processing/Handlers/IMotionCaptureHandler.cs
```diff
@@ -33,6 +33,7 @@ public interface IMotionCaptureHandler
         /// <summary>
         /// Disables motion detection. When configured, this will instruct the capture handler not to detect motion.
         /// </summary>
-        void DisableMotionDetection();
+        /// <param name="disableCallbackOnly">When true, motion detection will continue but the OnDetect callback will not be invoked.</param>
+        void DisableMotionDetection(bool disableCallbackOnly = false);
```
techyian, line 38: If a user calls this method and wants to re-enable the callback, how do they do this? Am I right in thinking they just modify `OnDetectEnabled` against a `FrameDiffDriver` instance?

src/MMALSharp.Processing/Handlers/FrameBufferCaptureHandler.cs
```diff
@@ -45,6 +44,19 @@ public FrameBufferCaptureHandler()
             : base()
         { }
 
+        /// <summary>
+        /// Creates a new <see cref="FrameBufferCaptureHandler"/> configured for motion detection using a raw video stream
+        /// where MMALStandalone.Instance is used (such as processing a pre-recorded file) rather than camera-based processing.
+        /// </summary>
+        /// <param name="motionConfig">The motion configuration.</param>
+        /// <param name="onDetect">A callback for when motion is detected.</param>
+        public FrameBufferCaptureHandler(MotionConfig motionConfig, Action onDetect)
```
techyian, line 54: I'm assuming this additional constructor is used to replicate what `MMALCamera.WithMotionDetection()` does?

src/MMALSharp.Processing/Processors/Motion/FrameDiffDriver.cs
```diff
@@ -0,0 +1,311 @@
...
+
+        private Action _onDetect;
+        private bool _firstFrame = true;
+        private MotionConfig _motionConfig;
+        private bool _fullTestFrame;
+        private ImageContext _imageContext;
+        private Stopwatch _testFrameAge;
+
+        /// <summary>
+        /// Fully are skipped when comparing the test frame to the current frame.
```
techyian: Think this is a typo? Is it supposed to read "Frames" instead of "Fully"?

src/MMALSharp.Processing/Processors/Motion/FrameDiffDriver.cs
```diff
@@ -0,0 +1,311 @@
...
+
+        /// <summary>
+        /// Resets the state of the buffers so that a new test frame is
+        /// stored. Also resets any state in the motion detection algorithm.
+        /// </summary>
+        public void ResetAnalyser()
+        {
+            this.FullFrame = false;
+            _fullTestFrame = false;
+            this.WorkingData = new List<byte>();
```
techyian, line 138: Could we see any further benefit to moving `this.WorkingData` into a `byte[]` instead of a `List<byte>`?

techyian (commented):
This is looking really good Jon and when you've got time it would be really helpful to get some documentation together on how all this works and some typical use-case examples (in addition to the example provided?) - it'll also help me do some testing :)

src/MMALSharp.Processing/Handlers/IMotionCaptureHandler.cs
MV10: Calling `EnableMotionDetection` will re-enable the callback. I will add that to the parameter docs.

src/MMALSharp.Processing/Handlers/FrameBufferCaptureHandler.cs
MV10: By funny coincidence I was just looking at that myself, and I don't think it is needed. There was a point where I was trying to avoid modifying `MotionConfig` and I think it's left over from that. I will straighten it out.

src/MMALSharp.Processing/Processors/Motion/FrameDiffDriver.cs
MV10: It should say "Fully black pixels", not sure how that happened. It is fixed in the changes I haven't committed yet (more about that momentarily).

src/MMALSharp.Processing/Processors/Motion/FrameDiffDriver.cs
MV10: No, internally `List<byte>` is just an array, so we'd just end up doing the same work (resizing etc). One thing that might help, though, is to think about initializing the list to a certain size. I believe the default is just 10 entries and there is some sort of scale-up algorithm, it may be helpful to set it to a large size up front, given the amount of data we're working with. I'll open an issue so we remember to consider it later on. I think there are multiple places in the code where that idea might apply.

MV10:
Ian, thanks for taking time to look. I was sort of waiting for your feedback about changing `ImageContext` properties to fields. The benefits are significant, but that's a deviation from your general policy for public scope members. If you're OK with it based on the test results, I'll commit the final round of changes after looking at that constructor you mentioned above.

And yessir, I definitely agree, I have quite a lot of documentation work lined up!

MV10:
Oh and I'm also still unsure of how to set up the camera to get one raw frame into the handler -- it exits before the image processor is able to finish running. See this comment:

https://github.com/techyian/MMALSharp/issues/108#issuecomment-689754922

src/MMALSharp.Processing/Handlers/FrameBufferCaptureHandler.cs
MV10: Actually that constructor was useful when doing nothing but visualizing the motion detection analysis. Used in a simple configuration like this:

```csharp
using (var shell = new ExternalProcessCaptureHandler(raw_to_mjpeg_stream))
using (var motion = new FrameBufferCaptureHandler(motionConfig, null))
{
    motionAlgorithm.EnableAnalysis(shell);

    resizer.ConfigureOutputPort<VideoPort>(0, new MMALPortConfig(MMALEncoding.RGB24, MMALEncoding.RGB24, width: 640, height: 480), motion);
    cam.Camera.VideoPort.ConnectTo(resizer);

    await cameraWarmupDelay(cam);
    var timeout = new CancellationTokenSource(TimeSpan.FromSeconds(seconds));
    await Task.WhenAll(new Task[]{
            shell.ProcessExternalAsync(timeout.Token),
            cam.ProcessAsync(cam.Camera.VideoPort, timeout.Token),
        }).ConfigureAwait(false);
}

```

MV10: Oh I see I do explain one usage in the constructor summary, although it could be better, it does now also work with the camera (the example above is for streaming). I'll update the summary.

> Creates a new <see cref="FrameBufferCaptureHandler"/> configured for motion detection using a raw video stream where MMALStandalone.Instance is used (such as processing a pre-recorded file) rather than camera-based processing.

MV10:
Figured it was silly not to just merge the convolution changes, too. Everything is easy enough to roll back if necessary.

(The merge conflict was just around some stuff you'd changed in `ConvolutionBase` -- which is almost completely changed in my commit.)

src/MMALSharp.Processing/Processors/Effects/ConvolutionBase.cs
```diff
@@ -27,165 +52,76 @@ public abstract class ConvolutionBase
...
-                if (!context.Raw)
-                {
-                    using (var ms2 = new MemoryStream())
-                    {
-                        bmp.Save(ms2, context.StoreFormat);
-                        store = new byte[ms2.Length];
-                        Array.Copy(ms2.ToArray(), 0, store, 0, ms2.Length);
-                    }
-                }
+                throw new Exception("Convolution effects require raw frame data");
```
techyian: This concerns me a bit. I'm quite sure that image convolution worked on encoded images before - if they were encoded they were loaded into a new `Bitmap` instance and the metadata was made available via `BitmapData`. Is that not going to be possible in this re-write? Ideally it should be able to handle both.

techyian (commented):
The changes are looking good Jon. I've made a comment regarding needing images to be raw in order for the re-write to work properly.

src/MMALSharp.Processing/Processors/Effects/ConvolutionBase.cs
MV10: I think you're right, I think an optional extra step to load it into a new bitmap should work. I'll take a look at it from that perspective. It does ultimately need to be raw, though, so that `FrameAnalyser` can derive the bits (and bytes) per pixel. I don't think the in-memory bitmap conversion is too expensive. (I sort of wish we had a good abstraction for turning performance data on and off... maybe something for the future.)

Speaking of bytes per pixel, I'm still struggling to get `TakeRawPicture` to work. When I try the code shown below, `ImageContext.PixelFormat` is not set which causes `FrameAnalyser` to fail in the `GetBytesPerPixel` method. Am I overlooking something?

```csharp
var cam = GetConfiguredCamera();
MMALCameraConfig.Encoding = MMALEncoding.RGB24;
MMALCameraConfig.EncodingSubFormat = MMALEncoding.RGB24;
using (var imgCaptureHandler = new ImageStreamCaptureHandler(ramdiskPath, "jpg"))
{
    imgCaptureHandler.Manipulate(context =>
    {
        context.Apply(new SharpenProcessor());
    }, ImageFormat.Jpeg);
    await cam.TakeRawPicture(imgCaptureHandler);
}
cam.Cleanup();
```

Ah -- looking at the debug log, it looks like `StreamCaptureHandler` is involved. In the log I see `Applying image processing` followed by `Something went wrong while processing stream` and the null ref throw. (It was initially confusing because the handler only logs the inner exception -- there was none -- and not the actual exception!) But I'm not sure where the context comes from at that point in time.

MV10: Let me run my thinking by you:

Since my revisions operate on `ImageContext` data, do you know if copying data out of a _formatted_ `Bitmap` (sourced from a JPEG memory stream for example) yields pixel level data? I was trying to figure out if I need to put it through a format conversion to get raw data. The old version operated directly on the locked bitmap, I'm wondering how much voodoo the `Bitmap` object does to make the pixels of a lossy compressed image available (once you unlock it, I assume it's back to lossy JPEG again). I know when a bitmap region is locked, .NET actually makes another copy of that region and pins it in memory, so I'm guessing that it probably is decompressed at that point. So I suspect copying that really will get uncompressed pixels.

Originally I was thinking after the context is updated (by copying the locked `Bitmap` data), the point above implies `ImageContext.Raw` should then be `true` -- but then I realized that's probably not correct. The raw flag only applies to the three RGB formats, and I suppose the copied bitmap could be something like YUV, does that sound right?

So if that's right, after the frame is processed, the raw-to-formatted conversion really shouldn't necessarily require raw input at all, either. 

Do I sound like I'm on the right track here?

techyian: I was under the impression that the compression algorithm JPEG and other lossy formats use would remove full pixels, meaning the R, G, B and A if applicable would be removed. This means that you can still iterate over the data in the same way you would with raw data but there just wouldn't be as much of it. Once the data is lost by the lossy algorithm, the `Bitmap` class would have no way of retrieving that data again. Hopefully I'm right in my assumptions and apologies for taking so long to get back to you on this.

MV10: After doing a bunch of reading and more thinking about it, my question doesn't really make sense. JPEG is lossy but it isn't storing pixels nor is it "losing" pixels, what it's losing is more abstract (DCT compression is based on perceptual color models and frequency thresholds). So it would never really make sense to operate on the compressed version of an image, the only thing that data is useful for is to decompress back to a bitmap -- it isn't stored as pixels, in other words, it's just a data stream meant to feed into a decompression algorithm.

I was basically mixing up "raw" with "uncompressed" -- the "raw" formats are image data which has never been compressed, versus "uncompressed" meaning a full-pixel-count bitmap that was _stored_ as compressed. We'll have "all" of the pixels, but they will simply exhibit compression artifacts here and there.

I'm still hung up on `TakeRawPicture` not passing a correct `ImageContext`, even though I will get it working with non-raw images as well. I assume a raw image will perform better without the overhead of bitmap locking and copying.

I also have to apologize for taking so long to deal with this. Work is really running me ragged right now -- I'm doing weekends, evenings, etc. (spent about 20 hours last weekend starting [this](https://github.com/MV10/EventStreamDotNet/)).

techyian: Thanks for the clarification, I should have done more research before replying. Are you suggesting that the `Bitmap` class does in fact decompress the data when it gets initialised? Not sure why you're still having trouble, my changes in 8b07ab3 should have set the pixel format [here](https://github.com/techyian/MMALSharp/blob/8b07ab3b697ff1159d988db9a84e0501a7b331e6/src/MMALSharp/Components/MMALCameraComponent.cs#L340). I'll try and have a look today/tomorrow.

MV10: > Are you suggesting that the Bitmap class does in fact decompress the data when it gets initialised?

Exactly. When I said this:

> So it would never really make sense to operate on the compressed version of an image, the only thing that data is useful for is to decompress back to a bitmap -- it isn't stored as pixels, in other words, it's just a data stream meant to feed into a decompression algorithm.

I was essentially referring to the contents of a full `ImageContext` frame buffer when MMAL is outputting compressed data. I don't know what `Bitmap` does internally, maybe it's more efficient to store it compressed, but that doesn't matter to us. Once we lock it for read/write, the pinned copy is definitely decompressed to full resolution. The loss simply refers to image quality, not actual individual pixels.

About the `TakeRawPicture` problem and `ImageContext.PixelFormat` -- EDIT -- those changes aren't in my branch, I probably messed up the merge conflict resolution, I'll compare and clean it up. Yes, that fixes `PixelFormat` -- still outputs two frames though (maybe I'm misstating that, it invokes `Manipulate` twice), I thought you'd also changed that?

MV10: Ooooh... I forgot about another weird behavior I was seeing with `TakeRawPicture` -- it seems to terminate the program before the convolution processing actually finishes running. I mentioned that [here](https://github.com/techyian/MMALSharp/issues/108#issuecomment-689754922) after I added a bunch of output to trace line-by-line:

> I take that back. It's calling the delegate but it never completes. I guess that call is on another thread and the app shuts down before `ApplyConvolution` runs more than a couple of lines.

techyian: It definitely only gets called once for me, I'm using your branch with the latest changes. I did commit a change which prevents the Callback Handler from being called if the buffer size is empty which prevents `Manipulate` from being called twice. Not sure why you're still seeing that, are you modifying the resolution at all or just leaving everything except the encoding & pixel formats the same?

> Ooooh... I forgot about another weird behavior I was seeing with TakeRawPicture -- it seems to terminate the program before the convolution processing actually finishes running. I mentioned that here after I added a bunch of output to trace line-by-line:

Again, not something I can reproduce. I've put a log call at the end of the `ApplyConvolution` method and it's outputting to the log OK. Am I missing something here?

As an aside, I'm still seeing a blue hue when applying convolutions to a raw image and then saving the output to an encoded format via `Bitmap.Save()`. I'm wondering if the RGB indexes of the image are being operated on in the wrong order and are being stored as BGR instead? If I run the raw modified image through a `MMALImageEncoder` instead of allowing `Bitmap` to save then the colours are OK ~~but there are other issues that occur when I do this which I'll make separate tickets for~~. None of this is a regression based on your work from what I can see, although v0.6 appeared to support applying convolutions to raw images, this in fact doesn't appear to be the case and only ever worked with encoded formats.

MV10: I took Friday off from work, so fingers crossed, I'm going to tackle this bright and early tomorrow!

MV10: Here's a weird finding to start the day:

My `TakeRawPicture` test set the camera to 1296x972 as a side-effect of earlier motion-detection testing (it's a full-frame binning mode). To my surprise, a different resolution was reported in an exception in the debug log: `Resolution 1312x976 has no recommended cell counts.` After a bit of searching, it turns out that the driver bumps the buffer up to the nearest 32-byte value ([RPi forum discussion](https://www.raspberrypi.org/forums/viewtopic.php?t=62364&start=275#p522410)):

> Because of the way the HW works, the width needs to be expanded to the nearest 32 bytes. So 1296 gets rounded up to 1312. The amount of valid data is still the same.

Have you encountered this before? Given the "valid data" comment, I'm thinking the `Resolution` field in `ImageContext` is a bit of a misnomer. It sounds as if only the width is padded to 32 bytes, so perhaps what's needed is a new field, `BufferWidth`, and to somehow set `Resolution.Width` to the true horizontal pixel count (which we still need for correct processing).

Edit: If altering the meaning of `Resolution.Width` or the need to decide between that and a new `BufferWidth` has too many side-effects elsewhere, it's super easy to just change my table of resolutions and cell counts -- but I think for convolution processing we'd see unwanted side-effects by processing the always-black padding stripe at the right edge of the buffer. (This problem didn't impact my earlier perf tests since I was loading the same saved-to-disk raw image over and over.)

I can work around this for now just by testing with a resolution naturally aligned to 32 bytes. (And now I see the blue hue, too...)

(The exception is also the reason I was saying earlier that execution was somehow being interrupted: I had expected an exception to be reported in the console by my outer `Program.Main` catch-block then terminate the program, but `StreamCaptureHandler` catches it and logs it -- which means it fails silently if logging is disabled. I don't know why but it was also the cause of the multiple delegate invocations. This design might be something to reconsider since this had unexpected side-effects.)

MV10: I hope I don't bother you too much with my thinking-out-loud style of working through this stuff. 💡 

The output port's zero-length buffer check doesn't seem to work consistently. (My copy definitely has that change now.)

Even though it's writing a processed JPG file now (albeit a blue one!), in the debug log I saw an index out of range error from the parallel cell processing loop, so I disabled that and wrote a serial loop with logging, thinking I'd see some kind of off-by-1 problem. Instead, the still image frame is processed, then a zero-length buffer is passed to the algorithm during shutdown, and that's where it fails. I assume the exception is some sort of dealloc/processing race condition, but that aspect is a side-effect -- it shouldn't be processing a zero-length buffer in the first place. It looks like that third output buffer (you mentioned earlier that it allocates three) _does_ correctly skip processing which is even more puzzling.

I'm wondering if you can think of where in the pipeline this might go wrong. (EDIT - I think it's the second part of the condition wrapping the output port callback, specifically `(eos && !this.Trigger.Task.IsCompleted)` ... I understand why it does that but I'm not sure how to address it.)

The first line of the log snippet below is the end of the correct cell processing (now at 640 x 480, hence 1024 cells). Then you can see a zero-length EOS buffer, and you can see that image processing kicks off again. The last entry below is the first pass of another cell processing loop. (It gets to cell 64 then throws index out of range, but obviously it shouldn't even start processing.)

I was thinking I'd short-circuit the whole thing on zero-length further downstream (I'm thinking in `PostProcess`), then I realized that has a zero-length-stream check, so I logged that. Towards the end of the logs below, you can see that the stream has 6K of "something" in it, which is why I think this second invocation happens. Not sure how to reconcile that with the zero buffer length though.

```
2020-10-09 08:33:01.369 -04:00 [INF] cell index 1023
2020-10-09 08:33:01.468 -04:00 [DBG] Releasing buffer.
2020-10-09 08:33:01.469 -04:00 [DBG] vc.ril.camera:out:2(RGB3): End of stream. Signaling completion...
2020-10-09 08:33:01.472 -04:00 [DBG] vc.ril.camera:out:2(RGB3): In native StillPort output callback
2020-10-09 08:33:01.472 -04:00 [DBG] vc.ril.camera:out:2(RGB3): In native output callback
2020-10-09 08:33:01.473 -04:00 [DBG] 
 Buffer Header 
---------------- 
Length: 0 
Presentation Timestamp: -9223372036854775808 
Flags: 
MMAL_BUFFER_HEADER_FLAG_EOS 
MMAL_BUFFER_HEADER_FLAG_COMPLETEFRAME 
---------------- 

2020-10-09 08:33:01.473 -04:00 [DBG] In managed Output callback
2020-10-09 08:33:01.473 -04:00 [DBG] Getting data from buffer
2020-10-09 08:33:01.473 -04:00 [DBG] Attempting to process data.
2020-10-09 08:33:01.474 -04:00 [DBG] PostProcess stream length: 67729
2020-10-09 08:33:01.474 -04:00 [DBG] Applying image processing.
2020-10-09 08:33:01.474 -04:00 [DBG] Setting parameter MMAL_PARAMETER_CAPTURE
2020-10-09 08:33:01.476 -04:00 [INF] cell index 0
```

MV10: Conclusions:

* Probably the extra stream data logged above is a race condition (shutdown begins after hardware started frame 2).
* The output port can't fix that problem, so partial frame data needs to be rejected in `PostProcess`.
* `PostProcess` can't rely on `ImageContext.EOS` because it's always true for still-image capture (frame 1 is the last). 
* `PostProcess` must decide whether `ImageContext.Data` represents a full frame.
* Compression means a full frame can't be recognized based on `ImageContext.Data` length like we can for raw data.

Looking at the logs, I'm unclear about the difference between the `COMPLETEFRAME` (which isn't in the MMAL docs) and `FRAMEEND` buffer flags, but I'm wondering if one of these could be set on a new `ImageContext` field to help full-frame recognition further into the pipeline. (I think that would be in `PortCallbackHandler`?)

I mean, I see that `COMPLETEFRAME` is defined as both start and end flags (ah, in the docs that's just `FRAME`) but I guess I'm questioning why MMAL doesn't appear to produce that when a buffer really is both the start and end? From the logs I only see the end frame flag with this 640 x 480 full-frame buffer (still testing with raw, by the way, in case that matters):

```
 Buffer Header 
---------------- 
Length: 921600 
Presentation Timestamp: 5326474418 
Flags: 
MMAL_BUFFER_HEADER_FLAG_FRAME_END 
---------------- 
```

EDIT - I can fix the problem in `PostProcess` by checking whether `ImageContext.Data.Length` is zero before reading and processing the partial stream ... but I'm wondering if this would be more appropriate at the end of `PortCallbackHandler` where the `Callback` method invokes `PostProcess` at EOS? Seems to me it should not be invoked at all if buffer length is zero, right? Regardless -- now I can move on, whichever fix you feel is appropriate -- and assuming I'm not stumbling into some other issue with this proposed fix!

MV10: Now it works with raw or formatted contexts, but of course a formatted context adds conversion overhead. It's probably worth mentioning that in the wiki for performance-sensitive uses. I'll explain "local context" in a moment but:

640x480 raw input, formatted output:
```
2020-10-09 14:31:44.654 -04:00 [DBG] ApplyConvolution set local context: 0 ms
2020-10-09 14:31:44.765 -04:00 [DBG] ApplyConvolution processed 1024 cells: 109 ms
2020-10-09 14:31:44.860 -04:00 [DBG] ApplyConvolution copy/formatted the results: 95 ms
```

640 x 480 JPEG input, formatted output:
```
2020-10-09 14:35:30.124 -04:00 [DBG] ApplyConvolution set local context: 94 ms
2020-10-09 14:35:30.225 -04:00 [DBG] ApplyConvolution processed 1024 cells: 100 ms
2020-10-09 14:35:30.270 -04:00 [DBG] ApplyConvolution copy/formatted the results: 44 ms
```

It occurs to me that because of the `StoreFormat` output stage, requiring raw input may not have been so onerous in the first place, but the work is done so no point in going back now.

When the input context is not already raw, it uses the `Bitmap` class to decode the data, then converts the pixel format to one of the raw pixel formats via `Bitmap.Clone`. The data isn't _actually_ raw -- as we discussed earlier, raw means never-compressed -- but it's still a raw _format_ which lets other parts of the system that do require raw data to operate normally. That data gets copied into a new local context used for processing. (I got rid of the `ImageContext` formatting extension class and brought the necessary methods back into `ConvolutionBase`, as I think they're probably not generically useful.)

After processing, that local context data is saved back to the original context using the requested `StoreFormat`, or if formatting wasn't provided, if the source wasn't originally a raw format, it array-copies the local context data back to the source context. 

Whew. It sounds simple now, but it took some pacing around the house to get all of that straight!

For some reason, `ImageContext.Resolution` isn't set when using `TakePicture`, although `TakeRawPicture` does set it. For now I read it from `Bitmap` after it loads the data, so this isn't a problem, but perhaps something to look into.

Raw-based output is still blue-tinted, although using encoded input is not. Curious...

The `PostProcess` zero-length check described in the preceding message is still there, too, and I'd still like to understand whether it might be more appropriate in `PortCallbackHandler`.

New commit momentarily...

MV10: To save you time, a recap of today's work:
* Convolution works with raw or formatted inputs
* ~~Open question [here](https://github.com/techyian/MMALSharp/pull/175#discussion_r502351729) about how to deal with non-32-byte-aligned resolutions~~
* The blue-tint problem with raw input (encoding is wrong somewhere I suspect)
* Zero-length buffer/data check in `PostProcess` or `PortCallbackHandler`? Any side-effects?

Other items that might warrant later consideration:
* Should `StreamCaptureHandler` silently suppress exceptions? Re-throw if logging is off?
* `TakePicture` doesn't set `ImageContext.Resolution`

MV10: Earlier I had overlooked that the padded buffer also changed the vertical resolution -- turns out that's padded to 16-byte values. That means almost all resolutions have a raw buffer size that is larger than the image resolution. For convolution, this only impacts `TakeRawPicture` since a formatted image from `TakePicture` is digested by `Bitmap` which produces a buffer matching the image resolution. 

I committed an update that just adds cell sizes for the padded buffer dimensions. ~~I can't decide if it would be better for `ImageContext.Resolution` to match the camera resolution or the true padded buffer size. As far as I can tell, the only effect of matching camera resolution would be to ignore the padding, which is what you'd probably want anyway.~~ EDIT - I temporarily added a line to set the true resolution but it trashes the image -- the stride is off! So I think just supporting the raw buffer dimensions with default cell values is the way to go.

techyian: > My TakeRawPicture test set the camera to 1296x972 as a side-effect of earlier motion-detection testing (it's a full-frame binning mode). To my surprise, a different resolution was reported in an exception in the debug log: Resolution 1312x976 has no recommended cell counts. After a bit of searching, it turns out that the driver bumps the buffer up to the nearest 32-byte value (RPi forum discussion):

I might need to do some investigation to check the camera's resolution is being configured correctly. The configuring of the camera component can be found in `MMALCameraComponent`. I'm not sure if the driver aligns on your behalf - I explicitly align the width and height to 32 and 16 respectively via the `Resolution` class and this library is influenced by Raspistill/Raspivid in terms of configuring the camera. 

> I'm wondering if you can think of where in the pipeline this might go wrong. (EDIT - I think it's the second part of the condition wrapping the output port callback, specifically (eos && !this.Trigger.Task.IsCompleted) ... I understand why it does that but I'm not sure how to address it.)

Well spotted, yes I think that's the problem. Could this be the solution:

```csharp
var eos = bufferImpl.AssertProperty(MMALBufferProperties.MMAL_BUFFER_HEADER_FLAG_FRAME_END) ||
                  bufferImpl.AssertProperty(MMALBufferProperties.MMAL_BUFFER_HEADER_FLAG_EOS) ||
                  this.ComponentReference.ForceStopProcessing;

if ((bufferImpl.CheckState() && bufferImpl.Length > 0 && !eos && !failed && !this.Trigger.Task.IsCompleted) || (eos && !this.Trigger.Task.IsCompleted && bufferImpl.Length > 0))
{
    this.CallbackHandler.Callback(bufferImpl);
}
```

Basically just making sure that the buffer length is always greater than 0 if we want to call the Callback Handler?

> For some reason, ImageContext.Resolution isn't set when using TakePicture, although TakeRawPicture does set it. For now I read it from Bitmap after it loads the data, so this isn't a problem, but perhaps something to look into.

Hmm, it should do. What is happening here is when `ConfigureOutputPort` is called against the encoder component (the output port being an instance of `StillPort`), the configure stage will check if the `MMALPortConfig` has a resolution set against it, and if not (which is the default), it will look [here](https://github.com/techyian/MMALSharp/blob/dev/src/MMALSharp/Ports/Outputs/StillPort.cs#L28) for the resolution to set it to. When your `ImageContext` is populated, the `Resolution` property is taken from the working port which in your case will be the encoder output port. If you are certain that it's not being retrieved correctly then please raise a ticket.

> Raw-based output is still blue-tinted, although using encoded input is not. Curious...

I have a feeling the raw output is being sent as BGR instead of RGB, I haven't had chance to look into this but it warrants a ticket raising rather than blocking this PR.

> Should StreamCaptureHandler silently suppress exceptions? Re-throw if logging is off?

Could you tell me where the exception is being suppressed and I'll review?

MV10: You're right about the raw image and RGB vs BGR, I temporarily added something to the end of the cell processing method to swap the R/B bytes and that fixes the color problem. I don't know where to start looking for an explanation though. Hmm, I wonder if [this](https://www.raspberrypi.org/forums/viewtopic.php?t=262579) firmware issue is related? (Although I'm not clear about the context of that issue, it sounds a bit like it might be specific to HDMI.)

MV10: Hi Ian, my browser's view into this PR hadn't updated with your comments until just now, after I'd already been playing around with the BGR/RGB issue. See comments above about the firmware issue which sounds related. I will also open a ticket about this -- I just now wrote a stand-alone test that does nothing but send `TakeRawPicture` to disk, then reads it back and does `Bitmap.Save` -- no other manipulation -- and the resulting JPEG has the colors reversed. Definitely not related to this PR.

> I might need to do some investigation to check the camera's resolution is being configured correctly. The configuring of the camera component can be found in MMALCameraComponent. I'm not sure if the driver aligns on your behalf - I explicitly align the width and height to 32 and 16 respectively via the Resolution class and this library is influenced by Raspistill/Raspivid in terms of configuring the camera.

I think this one is no longer a problem now that I've added the padded "resolutions" to the cell table, too. It's just the nature of the raw buffer that we have to work with it that way, I don't think there's benefit in messing with it further. About the only thing that _might_ make sense is to reflect the true resolution on `ImageContext` so that we don't waste time processing blank padding bytes, but honestly I think the overhead is immeasurably small.

>  Could this be the solution: Basically just making sure that the buffer length is always greater than 0 if we want to call the Callback Handler?

That does indeed fix the problem earlier in the pipeline! Committing that change now.

I will open a ticket about `TakePicture` not setting the resolution, I went so far as to put a `Console.WriteLine` as the first line of `ApplyConvolution` and it's definitely 0,0.

> Should StreamCaptureHandler silently suppress exceptions? Re-throw if logging is off?
> Could you tell me where the exception is being suppressed and I'll review?

https://github.com/MV10/MMALSharp/blob/motiondiff/src/MMALSharp.Processing/Handlers/StreamCaptureHandler.cs#L77

MV10:
EDIT -- In case you read the original by email -- I realized I was thinking about that wrong (distracted by yesterday's BGR/RGB tests that used `Bitmap.Save` to a file, whereas convolution saves to a memory stream) -- but this really should be a stand-alone issue, so I'll rewrite it.

src/MMALSharp.Processing/Processors/Effects/ConvolutionBase.cs
MV10: Pushed another fix to copy the cloned bitmap resolution back to the original context (which would still be 0,0 when invoked from `TakePicture`) -- EDIT - this is really about an edge-case where you're going from a formatted picture to raw output with multiple passes:

```csharp
imgCaptureHandler.Manipulate(context =>
{
    context.Apply(new SharpenProcessor());
    context.Apply(new GaussianProcessor(GaussianMatrix.Matrix5x5));
}, null);
```

Earlier I badly misstated how/why this is needed, but without this, the scenario above would crash. (Ironically, you'd still get raw output because of the way the stream handler works, but it wouldn't have effects applied.)

techyian:
Hi Jon, I've merged this PR now so thank you for your efforts on this. It would be great when you get some spare time if you could produce some documentation to surround this work and some useful examples. Now that this is merged I will begin looking at the tickets created off the back of this PR.

Thanks,
Ian

MV10:
Fantastic! Yes, I will get to work on the wiki soon. Thanks for all the help and patience with this one.

MV10:
Quick-pass wiki updates -- convolutions (the low-hanging fruit for documenting this PR):

https://github.com/techyian/MMALSharp/wiki/Image-Processing

...and a note that motion detection changes a lot in v0.7: 

https://github.com/techyian/MMALSharp/wiki/Advanced-Examples#FrameDiffDetection

Finally, I started that new CCTV page, which I will start with motion detection, then get into still frame captures, the various options, real-time analysis output, etc. -- just a placeholder today, but I'll try to work on this a bit each afternoon:

https://github.com/techyian/MMALSharp/wiki/CCTV-and-Motion-Detection

Do you know if it's possible to store images in the wiki itself? I thought of a trick that might work -- open a new issue, paste the image there (which uploads it to github) then reference that URL from the wiki. The issue can be closed but new images can still be added later...

MV10:
Quite a bit of content added today to the motion/CCTV wiki... (had a work day that involved mostly waiting on other people!)

techyian:
Very impressed with the write up, Jon, so thank you for that. Although I've been following the work you've been doing lately, seeing it all written down has really helped me understand the full picture of what you've been working on. I've re-jigged the tickets in the  v0.7 milestone so I think once the motion stuff is completed and other minor bits, I'll look at pushing a new official release out. Great job.

techyian:
> Do you know if it's possible to store images in the wiki itself?

I think it would probably be easier to create a resources (or similarly named) folder containing all the images used in the wiki. Send a PR in if you want and then you can reference them in your docs.

MV10:
Thanks. I know I can be hard to follow when I'm in the thick of it, but I'm still really happy about finding your library and your willingness to consider some pretty major changes. 

I went ahead and finished it up and I did reference my one motion analysis image from this PR, which I will probably point to a youtube capture of motion analysis at some point. Longer term I think there might be some benefit in diagrams elsewhere in the wiki to help explain component relationships and that sort of thing (I'm trying to think about stuff I had some trouble understanding initially), so I'll open a ticket when I'm feeling like I'm in a documenting mood. (It comes and goes -- and after writing that monster CCTV section, it's very _gone_ right now!)

I have some other ideas to further improve motion detection but I think the Really Big Stuff is settled.

Have a good one, talk to you soon.

